PR #499, opened 2016-08-01
Initial commit for Instance Details Table

ammendonca:


ammendonca:
image

objectiser:
Looks good. Does it support pagination?

Does the text filtering work on any part of the property entries. I.e a part of the key separator and part of value?

ammendonca:
No pagination support yet, it has to be client-side, right ? Filtering is on anything on the object (properties, principal, host name, ip address, duration, timestamp, and others which are present in the response and not shown). This may be inconvenient as filtering by strings which are not shown may still show the results and result in a weird experience for the user.. what fields should we look at ?

BTW, the instance count is returning results only for the last hour.. so, although there is data, it shows 0 if not in the last hour (even if time range is for 1 week)

I'm thinking about moving forward to the Detailed Instance View and improve this later with further optimizations, wdyt ?

objectiser:
Filtering can be left for now and we'll see how it works out. 

Pagination will need to be client side for now so need a cap on the max entries returned. But need to make it clear only partial results displayed.

In terms of the 1 hour range for the count it sounds like a bug so create a jira but make sure the date range is past in the criteria. 

+1 to moving to instance view.

objectiser:
Forgot to say can you get pavol to try it out before merge. Thanks

pavolloffay:
@objectiser so is it good to go after my review?

objectiser:
Check with Alex whether adding client side pagination would be quick. If not then merge after your review.

ammendonca:
Added client-side pagination

image

pavolloffay:
I think that the location of the button for showing instances details would be better located nex to "Root node" dropdown. I think it is more intuitive as the left sidebar is for filtering, but this might be only my opinion. What do you think @objectiser ?

button-location

When I set duration from 0 to 0 it shows all instances. When it is set from 1 to 1 it shows only instances from 1 to 1.

hwk0-0

hwk1-1

ammendonca:
@pavolloffay I agree with the location you've suggested, it's more in context (and probably just a magnifier icon would be enough). As for the 0 to 0 issue, I've seen what's happening, will fix. Thanks!

@objectiser let me know if you agree with the location, to change it.

objectiser:
+1 to moving the location

pavolloffay:
@ammendonca Thanks for improvements. Can I merge it? Or do you want to add something more?

pavolloffay:
Script for generating test data.
[apm-generate-traces.sh.zip](https://github.com/hawkular/hawkular-apm/files/399305/apm-generate-traces.sh.zip)

PR #502, opened 2016-08-03
HWKAPM-520 | Set fault field based on HTTP status code in zipkin span

pavolloffay:
https://issues.jboss.org/browse/HWKAPM-520

### DIFF
--- a/processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/FragmentCompletionTimeDeriver.java
+++ b/processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/FragmentCompletionTimeDeriver.java
@@ -17,12 +17,15 @@
 package org.hawkular.apm.processor.zipkin;
 
 import java.net.URL;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import org.hawkular.apm.api.model.Constants;
+import org.hawkular.apm.api.model.Property;
 import org.hawkular.apm.api.model.events.CompletionTime;
+import org.hawkular.apm.server.api.model.zipkin.BinaryAnnotation;
 import org.hawkular.apm.server.api.model.zipkin.Span;
 import org.hawkular.apm.server.api.task.AbstractProcessor;
 import org.hawkular.apm.server.api.task.RetryAttemptException;
@@ -81,6 +84,8 @@ public class FragmentCompletionTimeDeriver extends AbstractProcessor<Span, Compl
 
             ct.setTimestamp(item.getTimestamp()/1000);
 
+            setFault(ct, item.getBinaryAnnotations());
+
             if (log.isLoggable(Level.FINEST)) {
                 log.finest("FragmentCompletionTimeDeriver ret=" + ct);
             }
@@ -88,6 +93,18 @@ public class FragmentCompletionTimeDeriver extends AbstractProcessor<Span, Compl
         }
 
         return null;
+    }
+
+    private void setFault(CompletionTime completionTime, List<BinaryAnnotation> binaryAnnotations) {
+        List<HttpCodesUtil.HttpCode> errorCodes =
+                HttpCodesUtil.getClientOrServerErrors(HttpCodesUtil.getHttpStatusCodes(binaryAnnotations));
+
+        if (errorCodes.size() > 0) {
+            HttpCodesUtil.HttpCode errorCode = errorCodes.iterator().next();
+
+            completionTime.setFault(errorCode.getDescription());
+            completionTime.getProperties().add(new Property("http.status_code", String.valueOf(errorCode.getCode())));
+        }
     }
 
 }
--- a/processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/HttpCodesUtil.java
+++ b/processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/HttpCodesUtil.java
@@ -1,0 +1,138 @@
+/*
+ * Copyright 2015-2016 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.hawkular.apm.processor.zipkin;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Locale;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+import org.apache.http.impl.EnglishReasonPhraseCatalog;
+import org.hawkular.apm.server.api.model.zipkin.BinaryAnnotation;
+
+/**
+ * @author Pavol Loffay
+ */
+public class HttpCodesUtil {
+
+    private static final Logger log = Logger.getLogger(HttpCodesUtil.class.getName());
+
+    protected static final String ZIPKIN_HTTP_CODE_KEY = "http.status_code";
+
+    private HttpCodesUtil() {}
+
+    /**
+     * Method returns list of http status codes.
+     *
+     * @param binaryAnnotations zipkin binary annotations
+     * @return http status codes
+     */
+    public static List<HttpCode> getHttpStatusCodes(List<BinaryAnnotation> binaryAnnotations) {
+        if (binaryAnnotations == null) {
+            return Collections.emptyList();
+        }
+
+        List<HttpCode> httpCodes = new ArrayList<>();
+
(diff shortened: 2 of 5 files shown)


pavolloffay:
@jpkrohling could you please review when you have some time?

processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/HttpCodesUtil.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private HttpCodesUtil() {}
+
+    /**
+     * Method returns list of http status codes.
+     *
+     * @param binaryAnnotations zipkin binary annotations
+     * @return http status codes
+     */
+    public static List<HttpCode> getHttpStatusCode(List<BinaryAnnotation> binaryAnnotations) {
```
jpkrohling: s/getHttpStatusCode/getHttpStatusCodes/ , plural

processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/FragmentCompletionTimeDeriver.java
```diff
@@ -90,4 +94,13 @@ public CompletionTime processOneToOne(String tenantId, Span item) throws RetryAt
         return null;
     }
 
+    private void setFault(CompletionTime completionTime, List<BinaryAnnotation> binaryAnnotations) {
+        List<HttpCodesUtil.HttpCode> errorCodes =
+                HttpCodesUtil.getClientOrServerErrors(HttpCodesUtil.getHttpStatusCode(binaryAnnotations));
+
+        if (errorCodes.size() > 0) {
+            completionTime.setFault(errorCodes.iterator().next().getDescription());
```
jpkrohling: Is there a reason to return only the first error description, instead of joining them all together?

processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/NodeDetailsDeriver.java
```diff
@@ -86,10 +88,22 @@ public NodeDetails processOneToOne(String tenantId, Span item) throws RetryAttem
...
         return nd;
     }
 
+
+    private void setFault(NodeDetails nodeDetails, List<BinaryAnnotation> binaryAnnotations) {
+        List<HttpCodesUtil.HttpCode> errorCodes =
+                HttpCodesUtil.getClientOrServerErrors(HttpCodesUtil.getHttpStatusCode(binaryAnnotations));
+
+        if (errorCodes.size() > 0) {
+            nodeDetails.setFault(errorCodes.iterator().next().getDescription());
```
jpkrohling: Same comment as above.

processors/zipkin/src/test/java/org/hawkular/apm/processor/zipkin/HttpCodesUtilTest.java
```diff
@@ -0,0 +1,77 @@
...
+        BinaryAnnotation http400 = createHttpCodeBinaryAnnotation("400");
+        BinaryAnnotation http302 = createHttpCodeBinaryAnnotation("302");
+        BinaryAnnotation http500 = createHttpCodeBinaryAnnotation("500");
+
+        BinaryAnnotation other = new BinaryAnnotation();
+        other.setKey("return code");
+        other.setValue("450");
+
+        List<BinaryAnnotation> binaryAnnotationList = Arrays.asList(http200, http302, http400, http500, other);
+        Assert.assertEquals(4, HttpCodesUtil.getHttpStatusCode(binaryAnnotationList).size());
```
jpkrohling: If you also believe that explicit is better than implicit, then you'd want a special test/assertion for "return code", instead of adding it to a list and letting the reader figure out why/which should not be in the getHttpStatusCode list :)

On this case, an assertion that iterates the list and asserts that there's no "return code" on the list would suffice, IMO.

processors/zipkin/src/test/java/org/hawkular/apm/processor/zipkin/HttpCodesUtilTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+        binaryAnnotationList = Arrays.asList(other, other);
+        Assert.assertEquals(0, HttpCodesUtil.getHttpStatusCode(binaryAnnotationList).size());
```
jpkrohling: This *could* fix the comment above, but I'm left wondering why you created an array with two `other` there. What would happen if there were two `http200` instead of `other`? Is this what this is testing? :)

processors/zipkin/src/test/java/org/hawkular/apm/processor/zipkin/HttpCodesUtilTest.java
```diff
@@ -0,0 +1,77 @@
...
+        HttpCodesUtil.HttpCode code200 = new HttpCodesUtil.HttpCode(200, "Ok");
+        HttpCodesUtil.HttpCode code302 = new HttpCodesUtil.HttpCode(302, "Found");
+        HttpCodesUtil.HttpCode code400 = new HttpCodesUtil.HttpCode(400, "Bad Request");
+        HttpCodesUtil.HttpCode code501 = new HttpCodesUtil.HttpCode(501, "Not Implemented");
+
+        List<HttpCodesUtil.HttpCode> codeList = Arrays.asList(code200, code400, code400, code302, code501);
+
+        List<HttpCodesUtil.HttpCode> clientOrServerErrors = HttpCodesUtil.getClientOrServerErrors(codeList);
+        Assert.assertEquals(3, clientOrServerErrors.size());
+        Assert.assertEquals(Arrays.asList(code400, code400, code501), clientOrServerErrors);
```
jpkrohling: Is this assertion "order dependent"? If I change the order on `codeList`, would the test break? And more importantly: *should* the test break?

processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/HttpCodesUtil.java
pavolloffay: :+1:

jpkrohling:
Reviewed: my commented on the "outdated diffs" are still relevant for the latest commit.

processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/FragmentCompletionTimeDeriver.java
pavolloffay: I think there should be only one this type of BinaryAnnotation.

jpkrohling: Then the method should not return a list?

processors/zipkin/src/test/java/org/hawkular/apm/processor/zipkin/HttpCodesUtilTest.java
pavolloffay: I think the order matters, client should expect the result have same order as input binary annotations.

pavolloffay:
@jpkrohling thanks for the review. Updated code is here. If it's fine I will merge it.

processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/FragmentCompletionTimeDeriver.java
```diff
@@ -90,4 +95,16 @@ public CompletionTime processOneToOne(String tenantId, Span item) throws RetryAt
         return null;
     }
 
+    private void setFault(CompletionTime completionTime, List<BinaryAnnotation> binaryAnnotations) {
+        List<HttpCodesUtil.HttpCode> errorCodes =
+                HttpCodesUtil.getClientOrServerErrors(HttpCodesUtil.getHttpStatusCodes(binaryAnnotations));
+
+        if (errorCodes.size() > 0) {
```
jpkrohling: It's still not clear if you expect only one error code (and thus, the method should return a single object instead of a list), or if you expect a list of error codes. If a list is expected, you could join the messages.

jpkrohling:
Other than the question about single objects vs. list of objects, it looks good to me.

processors/zipkin/src/main/java/org/hawkular/apm/processor/zipkin/FragmentCompletionTimeDeriver.java
pavolloffay: I expect one error code, but I was thinking that utility class can be more generic and return all of status codes in the list of binary annotations. Or is it confusing?

pavolloffay: e.g. can one call return two http status codes?..

jpkrohling: One HTTP request can get only one HTTP response, and an HTTP response can have only one HTTP status code. It's a bit confusing to me, as I was under the impression that you might be dealing with more than one HTTP request here. If that's not the case at all, then you are having a code that is more complex than it should, for the sake of possible future need. I'd say that if you need that in the future, then you can refactor that in the future.

pavolloffay: I would like to leave it in this more generic form. It can be also refactored to get any type of binary annotations. We will see what is needed in the future.

jpkrohling: Alright, then it looks good to me.

PR #508, opened 2016-08-06
Initial commit for Instance Details Diagram View

ammendonca:
Adds a Details column to the Instance Details Table, which when clicked
shows a diagram depicting the nodes which compose the entry.


ammendonca:
image

objectiser:
Thanks for all the hard work on this Alexandre! I will test it out when back on Wednesday. 

Have a well deserved vacation.

objectiser:
Thanks @ammendonca it works great. There was only one minor issue which I've described in HWKAPM-560 - if I get a chance I'll see if I can fix.

objectiser:
Only other comment is that when displaying more complex diagrams, the height used for displaying the diagram may be limiting. So might need to think about whether navigating to a separate page, or possibly opening up the cell more, may be options.

PR #524, opened 2016-08-16
HWKAPM-542 | HWKAPM-541 | HWKAPM-571 |Span service - trace fragment and end to end trace

pavolloffay:
https://issues.jboss.org/browse/HWKAPM-542
https://issues.jboss.org/browse/HWKAPM-541
https://issues.jboss.org/browse/HWKAPM-576

### DIFF
--- a/server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
+++ b/server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
@@ -20,8 +20,12 @@ package org.hawkular.apm.server.elasticsearch;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.elasticsearch.action.admin.indices.refresh.RefreshRequestBuilder;
 import org.elasticsearch.action.bulk.BulkRequestBuilder;
@@ -35,7 +39,15 @@ import org.elasticsearch.index.query.QueryBuilder;
 import org.elasticsearch.index.query.QueryBuilders;
 import org.elasticsearch.indices.IndexMissingException;
 import org.elasticsearch.search.SearchHit;
+import org.hawkular.apm.api.model.Property;
+import org.hawkular.apm.api.model.trace.Component;
+import org.hawkular.apm.api.model.trace.Consumer;
+import org.hawkular.apm.api.model.trace.InteractionNode;
+import org.hawkular.apm.api.model.trace.Node;
+import org.hawkular.apm.api.model.trace.Producer;
+import org.hawkular.apm.api.model.trace.Trace;
 import org.hawkular.apm.api.services.StoreException;
+import org.hawkular.apm.api.utils.NodeUtil;
 import org.hawkular.apm.server.api.model.zipkin.Span;
 import org.hawkular.apm.server.api.services.SpanService;
 import org.hawkular.apm.server.elasticsearch.log.MsgLogger;
@@ -118,7 +130,10 @@ public class SpanServiceElasticsearch implements SpanService {
 
             for (SearchHit searchHitFields : response.getHits()) {
                 try {
-                    spans.add(deserialize(searchHitFields.getSourceAsString(), Span.class));
+                    Span span = deserialize(searchHitFields.getSourceAsString(), Span.class);
+                    if (!span.serverSpan()) {
+                        spans.add(span);
+                    }
                 } catch (IOException ex) {
                     log.errorFailedToParse(ex);
                 }
@@ -190,6 +205,145 @@ public class SpanServiceElasticsearch implements SpanService {
 
             client.clear(tenantId);
         }
+    }
+
+    @Override
+    public Trace getTraceFragment(String tenantId, String id) {
+        Span span = getSpan(tenantId, id);
+
+        if (span == null) {
+            return null;
+        }
+
+        InteractionNode interactionNode = spanToNode(span);
+        interactionNode.setNodes(recursiveTraceFragment(tenantId, span));
+
+        Trace trace = spanToTrace(span);
+        trace.getNodes().add(interactionNode);
+
+        return trace;
+    }
+
+    @Override
+    public Trace getTrace(String tenantId, String id) {
+        Trace trace = getTraceFragment(tenantId, id);
+
+        if (trace != null) {
+            processConnectedFragment(tenantId, trace, trace, null);
+        }
+
+        return trace;
+    }
+
+    /**
+     * This method aggregates enhances the root trace, representing the
+     * complete end to end instance view, with the details available from
+     * a linked trace fragment that should be attached to the supplied
+     * Producer node. If the producer node is null then don't merge
+     * (as fragment is root) and just recursively process the fragment
+     * to identify additional linked fragments.
+     *
+     * @param tenantId The tenant id
+     * @param root The root trace
+     * @param fragment The fragment to be processed
+     * @param producer The producer node, or null if processing the top level fragment
+     */
+    protected void processConnectedFragment(String tenantId, Trace root, Trace fragment, Producer producer) {
+        if (producer != null) {
+            // Merge the properties associated with the fragment into the root
+            root.getProperties().addAll(fragment.getProperties());
+
+            // Attach the fragment root nodes to the producer
+            producer.getNodes().addAll(fragment.getNodes());
+        }
+
+        List<Producer> producers = new ArrayList<>();
+        NodeUtil.findNodes(fragment.getNodes(), Producer.class, producers);
+
+        for (Producer p : producers) {
+            if (!p.getCorrelationIds().isEmpty()) {
+                List<Trace> fragments = getTraceFragments(tenantId, p.getCorrelationIds().stream()
+                        .map(correlationIdentifier -> correlationIdentifier.getValue())
+                        .collect(Collectors.toList()));
+
+                for (Trace tf : fragments) {
+                    processConnectedFragment(tenantId, root, tf, p);
+                }
+            }
+        }
+    }
+
+    private List<Trace> getTraceFragments(String tenantId, List<String> ids) {
+        List<Trace> traces = new ArrayList<>();
+
+        for (String id: ids) {
+            Trace traceFragment = getTraceFragment(tenantId, id);
+            if (traceFragment != null) {
+                traces.add(traceFragment);
+            }
+        }
+
+        return traces;
+    }
+
+    private List<Node> recursiveTraceFragment(String tenantId, Span parent) {
+        List<Span> spanChildren = getChildren(tenantId, parent.getId());
+
+        if (spanChildren == null) {
+            return Collections.emptyList();
+        }
+
+        List<Node> nodes = new ArrayList<>();
+
+        for (Span child: spanChildren) {
+            InteractionNode node = spanToNode(child);
+
+            if (!parent.clientSpan()) {
+                nodes.add(node);
+            }
+
+            if (!child.clientSpan()) {
+                node.setNodes(recursiveTraceFragment(tenantId, child));
+            }
+        }
+
+        return nodes;
+    }
+
+    private Trace spanToTrace(Span span) {
+        if (span == null) {
+            throw new NullPointerException();
+        }
+
+        List<Property> spanProperties = span.properties();
+
+        Trace trace = new Trace();
+        trace.setId(span.getId());
+        trace.setStartTime(span.getTimestamp());
+        trace.setHostAddress(Span.ipv4Address(spanProperties));
+        trace.setProperties(new HashSet<>(span.properties()));
+
+        return trace;
+    }
+
+    private InteractionNode spanToNode(Span span) {
+        String endpointType = "HTTP";
+        String url = span.url() != null ? span.url().getPath() : null;
+
+        InteractionNode node;
+        if (span.serverSpan()) {
+            node = new Consumer(url, endpointType);
+            node.addInteractionId(span.getId());
+        }else if (span.clientSpan()) {
+            node = new Producer(url, endpointType);
+            node.addInteractionId(span.getId());
+        } else {
+            node = new Component(url, null);
+        }
+
+        node.setBaseTime(TimeUnit.NANOSECONDS.convert(span.getTimestamp(), TimeUnit.MICROSECONDS));
+        node.setDuration(TimeUnit.NANOSECONDS.convert(span.getDuration(), TimeUnit.MICROSECONDS));
+        return node;
     }
 
     private <T> T deserialize(String json, Class<T> type) throws IOException {
--- a/server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
+++ b/server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
@@ -17,11 +17,23 @@
 
 package org.hawkular.apm.server.elasticsearch;
 
+import static org.hamcrest.core.IsInstanceOf.instanceOf;
+
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
+import org.hawkular.apm.api.model.trace.Component;
+import org.hawkular.apm.api.model.trace.Consumer;
+import org.hawkular.apm.api.model.trace.InteractionNode;
+import org.hawkular.apm.api.model.trace.Node;
+import org.hawkular.apm.api.model.trace.NodeType;
+import org.hawkular.apm.api.model.trace.Producer;
+import org.hawkular.apm.api.model.trace.Trace;
 import org.hawkular.apm.api.services.StoreException;
 import org.hawkular.apm.server.api.model.zipkin.Annotation;
 import org.hawkular.apm.server.api.model.zipkin.AnnotationType;
@@ -29,6 +41,7 @@ import org.hawkular.apm.server.api.model.zipkin.BinaryAnnotation;
 import org.hawkular.apm.server.api.model.zipkin.Endpoint;
 import org.hawkular.apm.server.api.model.zipkin.Span;
 import org.hawkular.apm.server.api.services.SpanService;
+import org.hawkular.apm.server.api.utils.SpanUniqueIdGenerator;
 import org.junit.After;
 import org.junit.Assert;
 import org.junit.Before;
@@ -81,7 +94,7 @@ public class SpanServiceElasticsearchTest {
         span.setAnnotations(Arrays.asList(annotation));
         span.setBinaryAnnotations(Arrays.asList(binaryAnnotation));
 
-        storeAndWait(null, span);
+        storeAndWait(null, Collections.singletonList(span));
         Span spanFromDb = spanService.getSpan(null, span.getId());
 
         assertSpansEquals(span, spanFromDb);
@@ -93,7 +106,7 @@ public class SpanServiceElasticsearchTest {
         parentSpan.setId("parent");
         parentSpan.setName("foo parent");
 
-        storeAndWait(null, parentSpan);
+        storeAndWait(null, Collections.singletonList(parentSpan));
 
         Annotation annotation = new Annotation();
         annotation.setValue("foo bar");
@@ -117,13 +130,13 @@ public class SpanServiceElasticsearchTest {
         span.setAnnotations(Arrays.asList(annotation));
         span.setBinaryAnnotations(Arrays.asList(binaryAnnotation));
 
-        storeAndWait(null, span);
+        storeAndWait(null, Collections.singletonList(span));
 
         span = new Span();
         span.setId("id2");
         span.setParentId("parent");
 
-        storeAndWait(null, span);
+        storeAndWait(null, Collections.singletonList(span));
 
         List<Span> children = spanService.getChildren(null, "parent");
 
@@ -137,15 +150,15 @@ public class SpanServiceElasticsearchTest {
         Span clientSpan = new Span();
         clientSpan.setId("foo");
         clientSpan.setDuration(111);
-        clientSpan.setAnnotations(annotations());
+        clientSpan.setAnnotations(clientAnnotations());
 
-        storeAndWait(null, clientSpan, x -> x.getId() + clientIdsuffix);
+        storeAndWait(null, Collections.singletonList(clientSpan), x -> x.getId() + clientIdsuffix);
 
         Span span = new Span();
         span.setId("foo");
         span.setDuration(444);
 
-        storeAndWait(null, span, x -> x.getId());
+        storeAndWait(null, Collections.singletonList(span), x -> x.getId());
 
         Span spanFromDb = spanService.getSpan(null, "foo");
         assertSpansEquals(span, spanFromDb);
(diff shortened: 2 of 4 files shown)


server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +198,61 @@ public void clear(String tenantId) {
...
+        List<Node> nodes = new ArrayList<>();
+
+        for (Span span: spanChildren) {
+            if (span.serverSpan()) {
+                continue;
+            }
+
+            // TODO set uri
+            Consumer consumer = new Consumer(null, "http");
+            consumer.addInteractionId(span.getId());
```
objectiser: Should only create a Consumer node if the span is a clientSpan(). Otherwise create a Component node type.

objectiser: The correlation id is correct. However there are other properties on the Consumer (and Component) nodes that will need to be set - including the baseTime and duration, endpointType (for Consumer) and componentType for Component - although this may have to wait until we have more configurable support.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +198,61 @@ public void clear(String tenantId) {
...
+        trace.setStartTime(span.getTimestamp());
+        trace.setHostAddress(Span.ipv4Address(spanProperties));
+        trace.setProperties(new HashSet<>(span.properties()));
+
+//        trace.setBusinessTransaction()
+//        trace.setHostName()
+//        trace.setPrincipal()
+//        trace.setNodes()
+
+        return trace;
```
objectiser: This should also result in the top level node being created - if the span is a serverSpan() then it should be represented by a Consumer, a clientSpan, then a Producer, and anytihng else would be a Component. So probably best to have this logic in one place.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
pavolloffay: Do I understand it correctly that It should not go deeper if it is a server span?

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +201,83 @@ public void clear(String tenantId) {
...
+        String endpointType = "http";
+        String url = span.url() != null ? span.url().toString() : null;
+
+        InteractionNode node;
+        if (span.serverSpan()) {
+            node = new Consumer(url, endpointType);
+        }else if (span.clientSpan()) {
+            node = new Producer(url, endpointType);
+        } else {
+            node = new Component(url, endpointType);
```
objectiser: Second parameter for Component is component type - which at the moment we don't know, so could just leave as null.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +201,83 @@ public void clear(String tenantId) {
...
+        }
+
+        node.setBaseTime(span.getTimestamp());
```
objectiser: baseTime and duration are in nano-seconds.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +201,83 @@ public void clear(String tenantId) {
...
+        /**
+         * TODO set duration ss - sr? or cr - cs which captures network latency?
+         */
+//        node.setDuration();
+        node.addInteractionId(span.getId());
```
objectiser: correlation id will only be relevant for Producer and Consumer nodes

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +201,83 @@ public void clear(String tenantId) {
...
+
+        LinkedList<Node> nodes = new LinkedList<>();
+
+        for (Span span: spanChildren) {
+            if (span.serverSpan()) {
+                continue;
+            }
+
+            InteractionNode node = spanToNode(span);
+            node.setNodes(recursiveTraceFragment(tenantId, span.getId()));
```
objectiser: A better way to stop the recursion may be to check that the span is not a clienSpan(), instead of the continue on a serverSpan().

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +201,83 @@ public void clear(String tenantId) {
         }
     }
 
+    @Override
+    public Trace getTraceFragment(String tenantId, String id) {
+        Span span = getSpan(tenantId, id);
+
+        Trace trace = spanToTrace(span);
+        LinkedList<Node> nodes = recursiveTraceFragment(tenantId, id);
+        nodes.push(spanToNode(span));
```
objectiser: The result of 'spanToNode(span)' should be added as a node on the trace. The list 'nodes' should then be set as the node list on the node returned from spanToNode.

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,6 +179,58 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+        descendant2.setId("descendant2");
+        descendant2.setParentId("root");
+        descendant2.setTraceId("root");
+        descendant2.setAnnotations(clientAnnotations());
+
+        Span descendant3 = new Span();
+        descendant3.setId("descendant3");
+        descendant3.setParentId("root");
+        descendant3.setTraceId("root");
+        descendant3.setAnnotations(serverAnnotations());
```
objectiser: I don't think you could ever get a server span directly contained within another server span.

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,6 +179,58 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+
+        Span descendant4 = new Span();
+        descendant4.setId("descendant4");
```
objectiser: The client span would have the same id as the server span that sent the message - that is how they will be correlated.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
pavolloffay: Is it a problem if it is set for a Component?

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +202,85 @@ public void clear(String tenantId) {
...
+    private List<Node> recursiveTraceFragment(String tenantId, String spanId) {
+        List<Span> spanChildren = getChildren(tenantId, spanId);
+        if (spanChildren == null) {
+            return new LinkedList<>();
+        }
+
+        List<Node> nodes = new ArrayList<>();
+
+        for (Span span: spanChildren) {
+            if (span.serverSpan()) {
```
objectiser: Would still prefer to remove this, and instead guard the call to 'recursiveTraceFragment' below which a !span.clientSpan() - as the children of a client span will only be server spans, so it is a redundant query.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +202,85 @@ public void clear(String tenantId) {
...
+                continue;
+            }
+
+            InteractionNode node = spanToNode(span);
+            nodes.add(node);
+
+            node.setNodes(recursiveTraceFragment(tenantId, span.getId()));
```
objectiser: so just wrap this statement with a if (!span.clientSpan()) {

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +202,85 @@ public void clear(String tenantId) {
...
+//        TODO remove
+//        trace.setBusinessTransaction()
+//        trace.setHostName()
+//        trace.setPrincipal()
+
+        return trace;
+    }
+
+    private InteractionNode spanToNode(Span span) {
+        String endpointType = "http";
```
objectiser: Sorry, minor thing - can you make this in capitals for consistency with trace fragment data.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +202,85 @@ public void clear(String tenantId) {
...
+            node = new Producer(url, endpointType);
+        } else {
+            node = new Component(url, null);
+        }
+
+        node.setBaseTime(TimeUnit.NANOSECONDS.convert(span.getTimestamp(), TimeUnit.MICROSECONDS));
+        /**
+         * TODO set duration ss - sr? or cr - cs which captures network latency?
+         */
+//        node.setDuration();
```
objectiser: It should be between the two annotations in the same span, so either sr -> ss, or cs -> cr.

Not sure what should happen with a component, as I don't know how they record the time taken within a span that is not a client or server?

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
objectiser: It would be misleading and might cause problems later if querying based on correlation id. The Component can have correlation ids, but these are intended for business relevant ids extracted from message data, May need to revisit the model at some point to make this clearer.

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,6 +179,66 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+        descendant.setAnnotations(clientAnnotations());
+
+        Span descendant2 = new Span();
+        descendant2.setId("descendant2");
+        descendant2.setParentId("root");
+        descendant2.setTraceId("root");
+        descendant2.setAnnotations(clientAnnotations());
+
+        Span descendant3 = new Span();
+        descendant3.setId("descendant3");
```
objectiser: If this is the 'server span' associated with descendant2''s client span, then it should have the same id, parent id and trace id (I believe). Haven't actually checked the docs, but this is what all the example data shows.

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,6 +179,66 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+        descendant3.setParentId("descendant2");
+        descendant3.setTraceId("descendant2");
+        descendant3.setAnnotations(serverAnnotations());
+
+        Span descendant4 = new Span();
+        descendant4.setId("descendant4");
```
objectiser: This is a client span - but indicates it's parent is another client span, which shouldn't happen.

pavolloffay:

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +204,81 @@ public void clear(String tenantId) {
...
+
+        Trace trace = spanToTrace(span);
+        trace.getNodes().add(interactionNode);
+
+        return trace;
+    }
+
+
+    private List<Node> recursiveTraceFragment(String tenantId, Span parent) {
+        List<Span> spanChildren = getChildren(tenantId, SpanUniqueIdGenerator.toOriginal(parent));
```
objectiser: Not necessary needing to change, but not sure whether the 'toOriginal' util function is required. The only id that is modified would be for client spans, which we wouldn't be requesting children for. Was there a specific case you found that required this?

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +204,81 @@ public void clear(String tenantId) {
...
+            return Collections.emptyList();
+        }
+
+        List<Node> nodes = new ArrayList<>();
+
+        for (Span child: spanChildren) {
+            InteractionNode node = spanToNode(child);
+            nodes.add(node);
+
+            if (node.getType() == NodeType.Component) {
```
objectiser: This needs to check for "!child.clientSpan()" - both Consumer and Component node types can have child nodes. It is only client spans that cannot have children.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +204,81 @@ public void clear(String tenantId) {
...
+        trace.setStartTime(span.getTimestamp());
+        trace.setHostAddress(Span.ipv4Address(spanProperties));
+        trace.setProperties(new HashSet<>(span.properties()));
+
+        return trace;
+    }
+
+    private InteractionNode spanToNode(Span span) {
+        String endpointType = "HTTP";
+        String url = span.url() != null ? span.url().toString() : null;
```
objectiser: This needs to be span.url().getPath() instead of toString()

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,16 +186,210 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+
+        Span clientDescendant2 = new Span();
+        clientDescendant2.setId("descendant2");
+        clientDescendant2.setParentId("root");
+        clientDescendant2.setTraceId("root");
+        clientDescendant2.setAnnotations(clientAnnotations());
+
+        Span serverDescendantOfDescendant2 = new Span();
+        serverDescendantOfDescendant2.setId("descendant2");
+        serverDescendantOfDescendant2.setParentId("descendant2");
```
objectiser: This should be setParentId("root") - should be same value as the associated clientDescendant2

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,16 +186,210 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+        serverDescendantOfDescendant2.setTraceId("root");
+        serverDescendantOfDescendant2.setAnnotations(serverAnnotations());
+
+        Span serverDescendantOfDescendant = new Span();
+        serverDescendantOfDescendant.setId("descendant");
+        serverDescendantOfDescendant.setParentId("descendant");
```
objectiser: Same as above, the setParentId should be 'root' - so same value as in the clientDescendant.getParentId()

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,16 +186,210 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+    public void testGetTraceFragmentClientRoot() throws StoreException, InterruptedException {
+        Span rootClientSpan = new Span();
+        rootClientSpan.setId("root");
+        rootClientSpan.setTraceId("root");
+        rootClientSpan.setName("rootClientSpan");
+        rootClientSpan.setAnnotations(clientAnnotations());
+
+        Span serverSpan = new Span();
+        serverSpan.setId("root");
+        serverSpan.setParentId("root");
```
objectiser: Not 100% sure, but I believe as the parentId in the client was null, the value for the server span would also be null.

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,16 +186,210 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+        serverSpan.setTraceId("root");
+        serverSpan.setAnnotations(serverAnnotations());
+
+        Span serverSpanDescendantRoot = new Span();
```
objectiser: Unclear which this is a server span for? If the initial rootClientSpan, then that client span can only have one corresponding server span, which is 'serverSpan'.

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,16 +186,210 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+
+        Span descendant = new Span();
+        descendant.setId("descendant");
+        descendant.setParentId("root");
+        descendant.setTraceId("root");
+        descendant.setAnnotations(clientAnnotations());
+
+        Span descendantServer = new Span();
+        descendantServer.setId("descendant");
+        descendantServer.setParentId("descendant");
```
objectiser: This parent id would need to be the same as the 'descendant.getParentId()'.

Just realized that this would cause problems retrieving children from the cache and store - the results shouldn't return server spans, as these would not be children, but do have the same parent id as their client span equivalents. I'll create a jira to investigate this.

objectiser:
@pavolloffay Review complete - still a few issues, but getting there. Don't forget (unless you know of cases where it is not this way), the parent id of the related client and server spans have the same value.

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
pavolloffay: But this can be tested,, see next update

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
pavolloffay: what if the root is a client span?

pavolloffay:
Can I do https://issues.jboss.org/browse/HWKAPM-541 in this PR?

objectiser:
@pavolloffay sure, as they are tightly related.

In response to the question about root client span - if the root node is a client span, then it won't have any children.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +203,85 @@ public void clear(String tenantId) {
...
+        if (spanChildren == null) {
+            return Collections.emptyList();
+        }
+
+        List<Node> nodes = new ArrayList<>();
+
+        for (Span child: spanChildren) {
+            InteractionNode node = spanToNode(child);
+
+            if (!(parent.serverSpan() && node.getType() == NodeType.Consumer ||
```
objectiser: No, I think the logic needs to go into the getChildren methods to not return server spans - as there are other situations where the getChildren method will be used. So revert this back to just adding the node.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearch.java
```diff
@@ -192,6 +203,85 @@ public void clear(String tenantId) {
...
+                    parent.clientSpan() && node.getType() == NodeType.Producer)) {
+                nodes.add(node);
+            }
+
+            if (node.getType() == NodeType.Component) {
```
objectiser: This guard condition still needs to be changed! "if (!child.clientSpan()) {"

pavolloffay:
Why cannot have root client span children?

A -> B -> C

A starts with client span `foo`, B generates a response - server span for client span `foo`. But B also calls other service C so it would generate a client span `bar` with parent `foo`.

server/elasticsearch/src/test/java/org/hawkular/apm/server/elasticsearch/SpanServiceElasticsearchTest.java
```diff
@@ -173,16 +186,209 @@ public void testGetSpanFromEmptyDB() throws StoreException, InterruptedException
...
+
+        InteractionNode rootProducerNode = (InteractionNode) trace.getNodes().get(0);
+        Assert.assertThat(rootProducerNode, instanceOf(Producer.class));
+        Assert.assertEquals(1, rootProducerNode.getCorrelationIds().size());
+        Assert.assertEquals("root", rootProducerNode.getCorrelationIds().get(0).getValue());
+        Assert.assertEquals(1, rootProducerNode.getNodes().size());
+        Assert.assertEquals(new HashSet<>(Arrays.asList("root2")),
+                extractCorrelationIds(rootProducerNode.getNodes()));
+
+        InteractionNode descendantConsumerNode = (InteractionNode) rootProducerNode.getNodes().get(0);
```
objectiser: It shouldn't be possible for a node to be contained in a Producer (not in a trace fragment anyway), so something must have gone wrong here.

objectiser:
Yes, so the Trace fragment for A only has a Producer node - representing the client span. It has no children (in terms of the Trace fragment).

The fragment only represents the behaviour that happens within a service, not across the service boundaries. Only the end to end trace represents the nodes across service boundaries and will therefore have Producer nodes that contain children (i.e. Consumer nodes).

pavolloffay:
Finally it is clear to me :octocat: :+1:

pavolloffay:
@objectiser thanks for an explanation of trace model. I think now it should be correct.

pavolloffay:
https://github.com/smoketurner/dropwizard-zipkin/blob/master/zipkin-example/src/main/java/com/example/helloworld/resources/HelloWorldResource.java

`curl -ivX GET 'http://localhost:8888/fetch' `

apm-trace

I think there should be only two nodes right?

objectiser:
No, this looks correct - the initial two represent receiving the /fetch and the sending the /result request - and then the third symbol represents receiving the /result request.

pavolloffay:
review note issue https://issues.jboss.org/browse/HWKAPM-576 should be also fixed here

objectiser:
@pavolloffay LGTM - only minor issue with the duration units in the UI - I'll investigate and possibly create a separate jira. Thanks!

PR #526, opened 2016-08-17
HWKAPM-575 | Instance table display problem when long property field

pavolloffay:
https://issues.jboss.org/browse/HWKAPM-575

apm-instance-table-wrap

Review note: it the text in property span is too long it would still overflow

### DIFF
--- a/ui/src/main/scripts/plugins/e2e/html/details-modal.html
+++ b/ui/src/main/scripts/plugins/e2e/html/details-modal.html
@@ -67,7 +67,7 @@
           <tr dir-paginate-start="tData in filteredData = (timesData | filter:tableFilter | filter:durationRange | orderBy:sortKey:reverse) | itemsPerPage:numPerPage">
             <td>{{ tData.timestamp | date:'medium' }}</td>
             <!--<td>{{ tData.uri }}</td>-->
-            <td>
+            <td style="overflow: hidden">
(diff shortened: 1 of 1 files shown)


objectiser:
:+1: Great, and thanks to Karel for the help!

pavolloffay:
Possible improvement: I can cut the property text and show only what fits in the column and create a jira for showing a tooltip with the whole text.

objectiser:
Could you create an image to show an example of what you are thinking?

pavolloffay:
Problem with the current approach is that if the property text is longer than the column width it overflows.

hawkular apm

objectiser:
Agree with making the change :) Although will need some thought, as the properties are also currently subject to the text filter - if the full value is not provided, that will affect the filtering. But that could be investigated in the associated jira.

pavolloffay:
There will be full text provided, but only part of it will be shown (what fits in the column width). The full text will be shown when a user puts a cursor at the property.

objectiser:
:+1: great

pavolloffay:
I think it is fixed:

hawkular apm-xxx

objectiser:
:+1:

pavolloffay:
Jira for tooltip: https://issues.jboss.org/browse/HWKAPM-580

objectiser:
@pavolloffay LGTM, feel free to merge.

@ammendonca FYI

pavolloffay:
@objectiser thanks!

PR #543, opened 2016-08-26
Travis ubuntu no maven

pavolloffay:
ubuntu trusty without custom maven.


pavolloffay:
https://docs.travis-ci.com/user/trusty-ci-environment/#JVM only maven 3.1.1

PR #544, opened 2016-08-26
Caching node to the user's home

jpkrohling:


jpkrohling:
Second build, with `-Pcache`:
```
[INFO] Total time: 03:43 min
```

Third build, with `-Pcache,tests`:
```
[INFO] Total time: 06:09 min
```

I believe the `-Pcache` won't have any effects on CI, as CI's should start from scratch, so, it would always rebuild the UI. But for local development, this should also help.

jpkrohling:
retest this please

pavolloffay:
@jpkrohling is there way to tell Jenkins not to clean UI artifact? If it does not clean then maven-frontend-plugin does not download and install npm, bower..

jpkrohling:
The `-Pcache` should do exactly that, but for a CI environment, starting with a clean workspace is the common practice. We could try, though.

PR #602, opened 2016-09-28
HWKAPM-593 : Fix for large trace instance diagrams

ammendonca:
- Moved the instance details table below the DT diagram;
- Clicking on the details icon the modal is shown with single trace instance

### DIFF
--- a/ui/src/main/scripts/plugins/btm/less/btm.less
+++ b/ui/src/main/scripts/plugins/btm/less/btm.less
@@ -47,6 +47,10 @@ he .dropdown {
   }
 }
 
+.modal-xl {
+  width: 95%;
+}
+
 .hk-txn-item {
 
   .hk-summary {
--- a/ui/src/main/scripts/plugins/directives/instance-view-diagram/ts/instanceViewDiagramDirective.ts
+++ b/ui/src/main/scripts/plugins/directives/instance-view-diagram/ts/instanceViewDiagramDirective.ts
@@ -178,9 +178,9 @@ render.shapes().component = function(parent, bbox, node) {
           let label = '<div class="name" style="color: #eee;">' + uri + '</div>';
           if (theShape === 'circle') {
             label = '<div><i class="fa fa-share-alt" style="color: #ddd; font-size: 2.5em; margin: 0.2em;"></i></div>';
-          } else if (d.componentType === 'Database') {
+          } else if (d.componentType && d.componentType.toLowerCase() === 'database') {
             label = '<div><i class="fa fa-database" style="color: #eee"></i></div>';
-          } else if (d.componentType === 'EJB') {
+          } else if (d.componentType && d.componentType.toLowerCase() === 'ejb') {
             label = '<div><i class="fa fa-coffee" style="color: #eee"></i></div>';
           }
 
--- a/ui/src/main/scripts/plugins/e2e/html/details-modal.html
+++ b/ui/src/main/scripts/plugins/e2e/html/details-modal.html
@@ -2,12 +2,13 @@
   <button type="button" class="close" ng-click="close()">
     <span class="pficon pficon-close"></span>
   </button>
-  <h4 class="modal-title">Instance Details for {{ rootNode.uri }} ({{filteredData.length}} entries shown)</h4>
+  <h4 class="modal-title" ng-hide="selectedTx">Instance Details for {{ rootNode.uri }} ({{filteredData.length}} entries shown)</h4>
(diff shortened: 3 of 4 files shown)


ammendonca:
image

image

image

image

objectiser:
@ammendonca Looks good, but found a couple of issues:

When I selected the first endpoint, it had 10 instances, selected one and viewed it - all fine.

I then did a different invocation against a second endpoint - I then selected that other endpoint - the top half of the page updated as expected, but the table did not get cleared. The only thing that changed on the table was the endpoint - but there should have only be a single instance.

Thinking that the table should disappear when a new endpoint is selected?

Second issue was scrolling - because I tend to use two fingers on the pad to scroll, this is also the zooming action on the diagram. So now I need to make sure the mouse pointer is located outside the diagram region to be able to scroll down to see the whole table. Not sure what the best option is here, whether to only show fewer rows? Thoughts?

objectiser:
apm-large-diagrams

objectiser:
@ammendonca Not actually part of the original issue, but would it also be possible to make the nodes on the instance diagram use the same uri[operation] label structure as in the distributed tracing page diagram? If you want it done as a separate issue, let me know and I will create a jira.

ammendonca:
> I then did a different invocation against a second endpoint - I then selected that other endpoint - the top half of the page updated as expected, but the table did not get cleared. The only thing that changed on the table was the endpoint - but there should have only be a single instance.
>
>Thinking that the table should disappear when a new endpoint is selected?

Right, it should. I can fix that quickly.

>Second issue was scrolling - because I tend to use two fingers on the pad to scroll, this is also the zooming action on the diagram. So now I need to make sure the mouse pointer is located outside the diagram region to be able to scroll down to see the whole table. Not sure what the best option is here, whether to only show fewer rows? Thoughts?

I've faced the same _issue_ several times while testing, but I can't figure a good way to overcome that. Showing fewer rows could be a way, but in case of the screenshot you've shared, we wouldn't have much more space than for a couple of entries (with space for pagination, filtering, etc).

For this reason I'd still prefer just the modal (which when expanded would be as it is now, large and with much space for the diagram).. but depends mostly on how important you think it is to view the DT diagram and the instances information at the same location.

objectiser:
If we went with your suggestion, how would the user move from the expanded diagram view, and back to the collapsed table showing all of the instances?

ammendonca:
I'd say we could toggle it to show/hide with the icon we currently have for showing it.

objectiser:
Sounds good to me - @cardosogabriel is this approach ok for you?

ammendonca:
Full modal alternative:
hwkapm-593-alt1

Taller area, but not full modal:
hwkapm-593-alt2

objectiser:
@ammendonca My vote would be for the first 'full modal'. Would it also be possible to make the distributed tracing region full page height? Or is this difficult to calculate?

objectiser:
only issue is whether it is intuitive that they should press the details icon again to collapse the table.

ammendonca:
>only issue is whether it is intuitive that they should press the details icon again to collapse the table.

Right, we can adjust the icon to make it more obvious, change it to a "Hide" button or just add this "Hide" button to the top-right of the diagram.

> Would it also be possible to make the distributed tracing region full page height? Or is this difficult to calculate?

I can try, I think with SVG it's hard to achieve this

cardosogabriel:
From the modals, I also prefer the full modal approach.
Suggestions of improvements:
- Remove (393 entries shown) from the modal title, since when you are seeing only one detail this information is inconsistent. You can place this information above the table, on the right side, for example.
- When seeing only one instance, place the link "Back to full table (383 entries)" above the table, where the filters where. Screenshot here.
table

ammendonca:
Agree with @cardosogabriel. Will fix accordingly.

@cardosogabriel, can you make the Distributed Tracing to have full height instead of the fixed 500px (https://github.com/hawkular/hawkular-apm/blob/master/ui/src/main/scripts/plugins/e2e/html/e2e.html#L24) ? I have been able to almost achieve it, but `calc` is playing tricks on me. Basically making the `card-pf` having the full height and then children div having 100% seems to work.

objectiser:
@cardosogabriel looks good

ammendonca:
Updated to use the full modal and with @cardosogabriel suggestions.

Also improved checking for EJB/Database type of nodes, being case insensitive.

objectiser:
@ammendonca Works great, thanks!

objectiser:
@cardosogabriel Thanks!

PR #623, opened 2016-10-07
HWKAPM-594 : Add tooltip to Instance View nodes

ammendonca:
Initial implementation of tooltip for showing node properties.

Considered the option of using a togglable table but faced some issues with passing the node properties outside of the diagram.

In order to prevent large values to overflow the tooltip, I've made it break when it reaches the tooltip width, here's an example:

image

Might come up with a better solution, but for now and if has to be released, this is a first attempt.


objectiser:
Would it be possible to use a different colour for the component background?

ammendonca:
Sure.. something "reddish" ?

image

objectiser:
Good enough for now.

ammendonca:
Updated

PR #643, opened 2016-10-18
HWKAPM-638 | Standardize time units across APM to microseconds

pavolloffay:
https://issues.jboss.org/browse/HWKAPM-638

This PR does not deal with UI changes. UI changes will be fixed in https://issues.jboss.org/browse/HWKAPM-602

Review notes:
Please review this carefully, especially unit and integration tests. There are changes in expected values in asserts.

### DIFF
--- a/api/src/main/java/org/hawkular/apm/api/model/trace/Trace.java
+++ b/api/src/main/java/org/hawkular/apm/api/model/trace/Trace.java
@@ -21,7 +21,6 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
 
 import org.hawkular.apm.api.model.Property;
 
@@ -48,6 +47,9 @@ public class Trace {
     @JsonInclude
     private String id;
 
+    /**
+     * Start time in microseconds
+     */
     @JsonInclude
     private long startTime;
 
@@ -86,14 +88,14 @@ public class Trace {
     }
 
     /**
-     * @return the startTime
+     * @return the startTime in microseconds
      */
     public long getStartTime() {
         return startTime;
     }
 
     /**
-     * @param startTime the startTime to set
+     * @param startTime the startTime in microseconds
      * @return The trace
      */
     public Trace setStartTime(long startTime) {
@@ -242,7 +244,7 @@ public class Trace {
     /**
      * This method returns the end time of the trace fragment.
      *
-     * @return The end time (in milliseconds)
+     * @return The end time (in microseconds)
      */
     public long endTime() {
         return getStartTime() + calculateDuration();
@@ -251,25 +253,22 @@ public class Trace {
     /**
      * This method returns the duration of the trace fragment.
      *
-     * @return The duration (in milliseconds), or 0 if no nodes defined
+     * @return The duration (in microseconds), or 0 if no nodes defined
      */
     public long calculateDuration() {
         if (!nodes.isEmpty()) {
-            long endTimeNS = 0;
+            long endTime = 0;
 
             for (int i = 0; i < getNodes().size(); i++) {
                 Node node = getNodes().get(i);
-                long et = node.overallEndTime();
+                long nodeEndTime = node.overallEndTime();
 
-                if (et > endTimeNS) {
-                    endTimeNS = et;
+                if (nodeEndTime > endTime) {
+                    endTime = nodeEndTime;
                 }
             }
 
-            long elapsedTime = endTimeNS - getNodes().get(0).getBaseTime();
-
-            // Convert elapsed time to milliseconds
-            return TimeUnit.MILLISECONDS.convert(elapsedTime, TimeUnit.NANOSECONDS);
+            return endTime - getNodes().get(0).getBaseTime();
         }
 
         return 0L;
--- a/api/src/main/java/org/hawkular/apm/api/services/Criteria.java
+++ b/api/src/main/java/org/hawkular/apm/api/services/Criteria.java
@@ -34,7 +34,13 @@ public class Criteria {
 
     private static final Logger log = Logger.getLogger(Criteria.class.getName());
 
+    /**
+     * Start time in milliseconds
+     */
     private long startTime = 0L;
+    /**
+     * End time in milliseconds
+     */
     private long endTime = 0L;
     private String businessTransaction;
     private Set<PropertyCriteria> properties = new HashSet<PropertyCriteria>();
@@ -45,14 +51,8 @@ public class Criteria {
     private String principal;
     private String uri;
     private String operation;
-
-    private static int DEFAULT_RESPONSE_SIZE = 100000;
-
-    private static long DEFAULT_TIMEOUT = 10000L;
-
-    private long timeout = DEFAULT_TIMEOUT;
-
-    private int maxResponseSize = DEFAULT_RESPONSE_SIZE;
+    private long timeout = 10000;
+    private int maxResponseSize = 100000;
 
     /**
      * Default constructor.
@@ -81,14 +81,14 @@ public class Criteria {
     }
 
     /**
-     * @return the startTime, or 0 meaning 1 hours ago
+     * @return the startTime in milliseconds, or 0 meaning 1 hours ago
      */
     public long getStartTime() {
         return startTime;
     }
 
     /**
-     * @param startTime the startTime to set
+     * @param startTime the startTime in milliseconds
      * @return The criteria
      */
     public Criteria setStartTime(long startTime) {
@@ -115,14 +115,14 @@ public class Criteria {
     }
 
     /**
-     * @return the endTime, or 0 meaning 'current time'
+     * @return the endTime in milliseconds, or 0 meaning 'current time'
      */
     public long getEndTime() {
         return endTime;
     }
 
     /**
-     * @param endTime the endTime to set
+     * @param endTime the endTime in milliseconds
      * @return The criteria
      */
     public Criteria setEndTime(long endTime) {
--- a/server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/ElasticsearchUtil.java
+++ b/server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/ElasticsearchUtil.java
@@ -16,6 +16,8 @@
  */
 package org.hawkular.apm.server.elasticsearch;
 
+import java.util.concurrent.TimeUnit;
+
 import org.elasticsearch.index.query.BoolQueryBuilder;
 import org.elasticsearch.index.query.FilterBuilder;
 import org.elasticsearch.index.query.FilterBuilders;
@@ -47,11 +49,14 @@ public class ElasticsearchUtil {
      */
     public static BoolQueryBuilder buildQuery(Criteria criteria, String timeProperty,
             String businessTxnProperty, Class<?> targetClass) {
-        long startTime = criteria.calculateStartTime();
-        long endTime = criteria.calculateEndTime();
-
+        /**
+         * Internally all time units are stored in microseconds
+         * Criteria API accepts milliseconds, therefore range adjustment is needed
+         */
         BoolQueryBuilder query = QueryBuilders.boolQuery()
-                .must(QueryBuilders.rangeQuery(timeProperty).from(startTime).to(endTime));
+                .must(QueryBuilders.rangeQuery(timeProperty)
+                        .from(TimeUnit.MILLISECONDS.toMicros(criteria.calculateStartTime()))
+                        .to(TimeUnit.MILLISECONDS.toMicros(criteria.calculateEndTime())));
 
         if (criteria.getBusinessTransaction() != null
                 && !criteria.getBusinessTransaction().trim().isEmpty()) {
--- a/server/elasticsearch/src/main/resources/hawkular-apm-mapping.json
+++ b/server/elasticsearch/src/main/resources/hawkular-apm-mapping.json
@@ -17,11 +17,6 @@
             }],
             "date_detection" : false,
             "numeric_detection": false,
-            "_timestamp": {
-                "enabled": true,
-                "path": "startTime",
-                "format": "yyyy/MM/dd HH:mm:ss"
-            },
             "properties": {
                 "businessTransaction": {
                     "type": "string",
@@ -36,8 +31,7 @@
                     "index": "not_analyzed"
                 },
                 "startTime": {
-                    "type": "date",
-                    "format": "yyyy/MM/dd HH:mm:ss"
+                    "type": "long"
                 },
                 "nodes": {
                     "type": "nested",
@@ -98,11 +92,6 @@
             }],
             "date_detection" : false,
             "numeric_detection": false,
-            "_timestamp": {
-                "enabled": true,
-                "path": "timestamp",
-                "format": "yyyy/MM/dd HH:mm:ss"
-            },
             "properties": {
                 "businessTransaction": {
                     "type": "string",
@@ -119,8 +108,7 @@
                     "type": "long"
                 },
                 "timestamp": {
-                    "type": "date",
-                    "format": "yyyy/MM/dd HH:mm:ss"
+                    "type": "long"
                 },
                 "properties": {
                     "type": "nested",
@@ -157,11 +145,6 @@
             }],
             "date_detection" : false,
             "numeric_detection": false,
-            "_timestamp": {
-                "enabled": true,
-                "path": "timestamp",
-                "format": "yyyy/MM/dd HH:mm:ss"
-            },
             "properties": {
                 "type": {
                     "type": "string",
@@ -200,8 +183,7 @@
                     }
                 },
                 "timestamp": {
-                    "type": "date",
-                    "format": "yyyy/MM/dd HH:mm:ss"
+                    "type": "long"
                 },
                 "properties": {
                     "type": "nested",
@@ -238,11 +220,6 @@
             }],
             "date_detection" : false,
             "numeric_detection": false,
-            "_timestamp": {
-                "enabled": true,
-                "path": "timestamp",
-                "format": "yyyy/MM/dd HH:mm:ss"
-            },
             "properties": {
                 "businessTransaction": {
                     "type": "string",
@@ -256,8 +233,7 @@
                     "type": "long"
                 },
                 "timestamp": {
-                    "type": "date",
-                    "format": "yyyy/MM/dd HH:mm:ss"
+                    "type": "long"
                 },
                 "properties": {
                     "type": "nested",
@@ -294,11 +270,6 @@
             }],
             "date_detection" : false,
             "numeric_detection": false,
-            "_timestamp": {
-                "enabled": true,
-                "path": "timestamp",
-                "format": "yyyy/MM/dd HH:mm:ss"
-            },
             "properties": {
                 "businessTransaction": {
                     "type": "string",
@@ -312,8 +283,7 @@
                     "type": "long"
                 },
                 "timestamp": {
-                    "type": "date",
-                    "format": "yyyy/MM/dd HH:mm:ss"
+                    "type": "long"
                 },
                 "properties": {
                     "type": "nested",
@@ -350,11 +320,6 @@
             }],
             "date_detection": false,
             "numeric_detection": false,
-            "_timestamp": {
-                "enabled": true,
-                "path": "timestamp",
-                "format": "yyyy/MM/dd HH:mm:ss"
-            },
             "properties": {
                 "traceId": {
                     "type": "string",
@@ -373,8 +338,7 @@
                     "index": "not_analyzed"
                 },
                 "timestamp": {
-                    "type": "date",
-                    "format": "yyyy/MM/dd HH:mm:ss"
+                    "type": "long"
                 },
                 "duration": {
                     "type": "long"
@@ -388,8 +352,7 @@
                     "include_in_root": true,
                     "properties": {
                         "timestamp": {
-                            "type": "date",
-                            "format": "yyyy/MM/dd HH:mm:ss"
+                            "type": "long"
                         },
                         "value": {
                             "type": "string",
--- a/server/processors-zipkin/src/test/java/org/hawkular/apm/server/processor/zipkin/CompletionTimeProcessingDeriverTest.java
+++ b/server/processors-zipkin/src/test/java/org/hawkular/apm/server/processor/zipkin/CompletionTimeProcessingDeriverTest.java
@@ -62,7 +62,7 @@ public class CompletionTimeProcessingDeriverTest {
         Assert.assertNull(completionTimeProcessing.getCompletionTime());
 
         completionTimeProcessing = completionTimeProcessingDeriver.processOneToOne(null, completionTimeProcessing);
-        Assert.assertEquals(1, completionTimeProcessing.getCompletionTime().getDuration());
+        Assert.assertEquals(rootSpan.getDuration().longValue(), completionTimeProcessing.getCompletionTime().getDuration());
 
         completionTimeProcessing = completionTimeProcessingDeriver.processOneToOne(null, completionTimeProcessing);
         Assert.assertNull(completionTimeProcessing);
@@ -75,15 +75,15 @@ public class CompletionTimeProcessingDeriverTest {
         descendant.setId("descendant");
         descendant.setTraceId("trace");
         descendant.setTimestamp(1500L);
-        Mockito.when(spanCacheMock.getTrace(null, "trace")).thenReturn(
-                new HashSet<>(Arrays.asList(rootSpan, descendant)));
+        Mockito.when(spanCacheMock.getTrace(null, "trace"))
+                .thenReturn(new HashSet<>(Arrays.asList(rootSpan, descendant)));
 
         completionTimeProcessing = new CompletionTimeProcessing(rootSpan);
         completionTimeProcessing = completionTimeProcessingDeriver.processOneToOne(null, completionTimeProcessing);
         Assert.assertNull(completionTimeProcessing.getCompletionTime());
 
         completionTimeProcessing = completionTimeProcessingDeriver.processOneToOne(null, completionTimeProcessing);
-        Assert.assertEquals(2, completionTimeProcessing.getCompletionTime().getDuration());
+        Assert.assertEquals(2000, completionTimeProcessing.getCompletionTime().getDuration());
 
         completionTimeProcessing = completionTimeProcessingDeriver.processOneToOne(null, completionTimeProcessing);
         Assert.assertNull(completionTimeProcessing);
--- a/server/processors/src/main/java/org/hawkular/apm/server/processor/tracecompletiontime/TraceCompletionInformationProcessor.java
+++ b/server/processors/src/main/java/org/hawkular/apm/server/processor/tracecompletiontime/TraceCompletionInformationProcessor.java
@@ -17,6 +17,7 @@
 package org.hawkular.apm.server.processor.tracecompletiontime;
 
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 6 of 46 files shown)


api/src/main/java/org/hawkular/apm/api/model/trace/Trace.java
```diff
@@ -269,7 +271,7 @@ public long calculateDuration() {
             long elapsedTime = endTimeNS - getNodes().get(0).getBaseTime();
```
objectiser: Can you rename any local vars that have used NS - otherwise could get confusing.

api/src/main/java/org/hawkular/apm/api/model/trace/Trace.java
```diff
@@ -269,7 +271,7 @@ public long calculateDuration() {
...
 
             // Convert elapsed time to milliseconds
```
objectiser: This comment, and the return value comments in calculateDuration and endTime methods should be changed as saying milliseconds.

api/src/main/java/org/hawkular/apm/api/services/Criteria.java
```diff
@@ -139,9 +140,9 @@ public Criteria setEndTime(long endTime) {
      */
     public long calculateEndTime() {
         if (endTime == 0) {
-            return System.currentTimeMillis();
+            return TimeUnit.NANOSECONDS.toMicros(System.nanoTime());
```
objectiser: We can't use nanoTime as a replacement for currentTimeMillis - which is a true timestamp. As described in the nanoTime docs `This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. `.

api/src/main/java/org/hawkular/apm/api/services/Criteria.java
```diff
@@ -139,9 +140,9 @@ public Criteria setEndTime(long endTime) {
...
         } else if (endTime < 0) {
-            return System.currentTimeMillis() - endTime;
+            return TimeUnit.NANOSECONDS.toMicros(System.nanoTime()) - endTime;
```
objectiser: Same as above.

api/src/main/java/org/hawkular/apm/api/services/Criteria.java
pavolloffay: What are the alternatives?

pavolloffay: Brave uses this https://github.com/pavolloffay/brave/blob/f511bc645590c116fca8170a69def6c2ef4f4e58/brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java#L150-L150 however it allows to submit annotations with timestamp precision higher than milliseconds. It seems that in JVM the safest way is to use `currentTimeMillis()`, we will loose some accuracy in our agent, but other clients (zipkin/opentracing) could use better precision.

pavolloffay:
Review data set: 
[vertxMicroSec.txt](https://github.com/hawkular/hawkular-apm/files/538759/vertxMicroSec.txt)
[vertxOdlTimeUnits.txt](https://github.com/hawkular/hawkular-apm/files/538760/vertxOdlTimeUnits.txt)

`curl -u jdoe:password -X POST -H "Content-Type: application/json" -d @$1 http://localhost:9411/hawkular/apm/traces/fragments`




Screenshots for microseconds:

hawkular apm-microseconds
hawkular apm-microseconds-components
hawkular apm-microseconds-details

Screenshots for old time units:
hawkular apm-component-oldtimeunits
hawkular apm-details-oldtimeunits
hawkular apm-oldtimeunits

server/processors/src/test/java/org/hawkular/apm/server/processor/fragmentcompletiontime/FragmentCompletionTimeDeriverTest.java
```diff
@@ -69,7 +69,7 @@ public void testProcessSingle() {
         assertFalse(ct.isInternal());
         assertEquals(trace.getStartTime(), ct.getTimestamp());
         assertEquals(c.getUri(), ct.getUri());
-        assertEquals(200, ct.getDuration());
+        assertEquals(200000000, ct.getDuration());
```
objectiser: Would expect 200000 - possibly source data needs converting to microseconds?

server/processors/src/test/java/org/hawkular/apm/server/processor/tracecompletiontime/TraceCompletionInformationInitiatorTest.java
```diff
@@ -103,7 +103,7 @@ public void testProcessSingleConsumerWithNoInteractionIdNoProducers() {
         assertFalse(ci.getCompletionTime().isInternal());
         assertEquals(trace.getStartTime(), ci.getCompletionTime().getTimestamp());
         assertEquals(c.getUri(), ci.getCompletionTime().getUri());
-        assertEquals(200, ci.getCompletionTime().getDuration());
+        assertEquals(200000000, ci.getCompletionTime().getDuration());
```
objectiser: Check source data. And same for below.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/AnalyticsServiceITest.java
```diff
@@ -108,7 +109,7 @@ public void initTest() {
     public void testGetUnboundEndpoints() throws Exception {
         Trace trace1 = new Trace();
         trace1.setId("1");
-        trace1.setStartTime(System.currentTimeMillis() - 4000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: The 4000 would need to be converted to microseconds  - although not sure why it says "within last hour" - but currently it represented 4 seconds, so needs to be 4000000 to be the same. However if the test works, then possibly not an issue.

If a fix is applied, then needs to be applied multiple times in this file.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/AnalyticsServiceITest.java
```diff
@@ -1132,8 +1133,8 @@ public void testGetCompletionTimeMultiFragment() throws Exception {
         traces.add(trace1);
         traces.add(trace2);
 
-        assertEquals(1000, trace1.calculateDuration());
-        assertEquals(1500, trace2.calculateDuration());
+        assertEquals(1000000000, trace1.calculateDuration());
```
objectiser: Would expect 1000000 and 1500000 - so possibly source data?

tests/dist/src/test/java/org/hawkular/apm/tests/dist/AnalyticsServiceITest.java
```diff
@@ -1159,7 +1160,7 @@ public void testGetCompletionTimeMultiFragment() throws Exception {
         assertNotNull(stats);
         assertEquals(1, stats.size());
 
-        assertEquals(1750, stats.get(0).getAverage());
+        assertEquals(1749999999, stats.get(0).getAverage());
```
objectiser: Looks like source data again - something not converted to microseconds. Should really be 1750000.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
```diff
@@ -75,10 +75,10 @@ public void initTest() {
     public void testGetCompletionTimesSingleFragment() throws Exception {
         Trace trace1 = new Trace();
         trace1.setId("1");
-        trace1.setStartTime(System.currentTimeMillis() - 4000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: Same as in previous file (and other places in this file. If fix applied in previous file, then apply here.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
```diff
@@ -101,17 +101,17 @@ public void testGetCompletionTimesSingleFragment() throws Exception {
     public void testGetCompletionTimesTwoFragmentInteractionP2PSync() throws Exception {
         Trace trace1 = new Trace();
         trace1.setId("1_2ip2psync");
-        trace1.setStartTime(System.currentTimeMillis() - 60000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: Was 60000 - so should now be 60000000 - but is 4000.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
```diff
@@ -145,17 +145,17 @@ public void testGetCompletionTimesTwoFragmentInteractionP2PSync() throws Excepti
     public void testGetCompletionTimesTwoFragmentInteractionP2PAsync() throws Exception {
         Trace trace1 = new Trace();
         trace1.setId("1_2ip2pasync");
-        trace1.setStartTime(System.currentTimeMillis() - 60000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: 60000 should be 60000000 not 4000.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
```diff
@@ -199,17 +199,17 @@ public void testGetCompletionTimesThreeFragmentControlFlowP2PSync() throws Excep
     protected void testGetCompletionTimesThreeFragmentP2PSync(Scope scope, String suffix) throws Exception {
         Trace trace1 = new Trace();
         trace1.setId("1_"+suffix);
-        trace1.setStartTime(System.currentTimeMillis() - 60000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: Same here.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
```diff
@@ -269,17 +269,17 @@ public void testGetCompletionTimesThreeFragmentControlFlowP2PAsync() throws Exce
     protected void testGetCompletionTimesThreeFragmentInteractionP2PAsync(Scope scope, String suffix) throws Exception {
         Trace trace1 = new Trace();
         trace1.setId("1_"+suffix);
-        trace1.setStartTime(System.currentTimeMillis() - 60000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: Same here.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
```diff
@@ -339,17 +339,17 @@ public void testGetCompletionTimesThreeFragmentControlFlowMultiConsumerAsync() t
     protected void testGetCompletionTimesThreeFragmentMultiConsumerAsync(Scope scope, String suffix) throws Exception {
         Trace trace1 = new Trace();
         trace1.setId("1_"+suffix);
-        trace1.setStartTime(System.currentTimeMillis() - 60000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: Same here.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
```diff
@@ -405,18 +405,18 @@ public void testGetCompletionTimesThreeFragmentCausedBy() throws Exception {
 
         Trace trace1 = new Trace();
         trace1.setId("1_"+suffix);
-        trace1.setStartTime(System.currentTimeMillis() - 60000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: Should be 60000000.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceServiceITest.java
```diff
@@ -129,7 +129,7 @@ public void testStoreAndRetrieveSimpleTraceById() {
     public void testStoreAndRetrieveComplexTraceById() {
         Trace trace1 = new Trace();
         trace1.setId("1");
-        trace1.setStartTime(System.currentTimeMillis());
+        trace1.setStartTime(TimeUnit.NANOSECONDS.toMicros(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis())));
```
objectiser: Extra 'TimeUnit.NANOSECONDS.toMicros( ... )'?

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceServiceITest.java
```diff
@@ -210,7 +210,7 @@ public void testStoreAndRetrieveComplexTraceById() {
     public void testStoreAndQueryAll() {
         Trace trace1 = new Trace();
         trace1.setId("1");
-        trace1.setStartTime(System.currentTimeMillis() - 4000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000); // Within last hour
```
objectiser: Same issue with converting 4000 to microseconds - if change is applied to other files.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceServiceITest.java
```diff
@@ -341,10 +341,10 @@ public void testStoreAndQueryEndTimeExclude() {
 
         // Query stored trace
         Criteria criteria = new Criteria();
-        criteria.setEndTime(1500);
+        criteria.setEndTime(2);
```
objectiser: These look wrong - the end times are being set precisely - so shouldn't the updated value be 1500000?

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceServiceITest.java
```diff
@@ -341,10 +341,10 @@ public void testStoreAndQueryEndTimeExclude() {
...
         Wait.until(() -> traceService.searchFragments(null, criteria).size() == 1);
 
-        criteria.setEndTime(1100);
+        criteria.setEndTime(1);
```
objectiser: Same comment as above.

server/elasticsearch/src/main/resources/hawkular-apm-mapping.json
```diff
@@ -17,11 +17,6 @@
             }],
             "date_detection" : false,
             "numeric_detection": false,
-            "_timestamp": {
```
pavolloffay: Why is it required?

server/processors/src/main/java/org/hawkular/apm/server/processor/tracecompletiontime/TraceCompletionInformationProcessor.java
```diff
@@ -177,9 +178,10 @@ protected void processCommunication(String tenantId, TraceCompletionInformation
             // TODO: Need to derive expiration based on knowledge about the time
             // different between the source and target linked fragments being
             // reported
-            long baseTimeStamp = ob.isMultiConsumer() ? cd.getTimestamp() : System.currentTimeMillis();
-            newc.setExpire(baseTimeStamp
-                        + TraceCompletionInformation.Communication.DEFAULT_EXPIRY_WINDOW);
+            long baseTimeStamp = TimeUnit.MICROSECONDS.toMillis(ob.isMultiConsumer() ?
```
pavolloffay: Can be, it produces the same result, but it's cleaner, so changing :+1:

server/elasticsearch/src/main/resources/hawkular-apm-mapping.json
objectiser: I think otherwise ES uses the time the document was recorded as a timestamp for query purposes.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/AnalyticsServiceITest.java
pavolloffay: Well its your code (`// Within last hour`). I think you did it because then you can query data with default criteria (returns data for last hour)

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
pavolloffay: why? it does not matter, you were using it to be able to use default query (better is to use the same constant across all tests, otherwise it looks suspicious)

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
pavolloffay: why? it does not matter, you were using it to be able to use default query (better is to use the same constant across all tests, otherwise it looks suspicious)

tests/dist/src/test/java/org/hawkular/apm/tests/dist/AnalyticsServiceITest.java
objectiser: Yes know it was my comment, but don't know what I was trying to say :) Possibly was just setting a value to override the default 1hour - but either way any time data needs to be converted to microseconds to be consistent with overall change.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
objectiser: Just from a review perspective, it makes it easier if there is a simple correlation between the pre/post change, as this is a large PR. Changing the values (other than from the perspective of units) means the review is more involved, to determine if the individual tests still make sense. But if you are sure they represent the same conditions, then it is not a major issue.

pavolloffay:
I'm not sure whether we should use `System.nanoTime()` for setting trace/span timestamp see http://docs.oracle.com/javase/7/docs/api/java/lang/System.html#nanoTime()

`This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative). The same origin is used by all invocations of this method in an instance of a Java virtual machine; other virtual machine instances are likely to use a different origin.

This method provides nanosecond precision, but not necessarily nanosecond resolution (that is, how frequently the value changes) - no guarantees are made except that the resolution is at least as good as that of currentTimeMillis().`

pavolloffay:
Btw. brave uses `System.currentTimeMillis()` to set span's timestamp.

objectiser:
That is the way in which it is being used - to measure the elapsed time between nodes being recorded.

The relative difference between the nanoTime recorded on child nodes, against the top level node, is then used as an elapsed time to add to the millisecond timestamp associated with the trace fragment - thus giving a nanosecond accurate timestamp for each node.

objectiser:
That is because brave is recording one span at a time - we have a reference point (i.e. the trace fragment being created) upon which we can measure the nanosecond elapsed time of the subsequent nodes that are added to the trace fragment.

server/elasticsearch/src/main/resources/hawkular-apm-mapping.json
pavolloffay: In our time ranged queries we use this code https://github.com/hawkular/hawkular-apm/blob/master/server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/ElasticsearchUtil.java#L53

Here is the doc for it: 
https://www.elastic.co/guide/en/elasticsearch/reference/1.7/mapping-timestamp-field.html
I'm not if it would work, since we store it as microseconds.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceCompletionTimeITest.java
pavolloffay: I have extracted that magic number to constant with a comment what is it for.

server/elasticsearch/src/main/resources/hawkular-apm-mapping.json
pavolloffay: It should be ms https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-timestamp-field.html and we set it to microseconds, so probably don't use it.

objectiser: Can you create a separate jira for this, as I think it can be addressed by using a custom serializer to essentially store an internal field in ES which could be referenced by the index mapping. As otherwise the timestamps won't be accurate.

server/processors-zipkin/src/test/java/org/hawkular/apm/server/processor/zipkin/CompletionTimeProcessingDeriverTest.java
```diff
@@ -75,15 +75,15 @@ public void testSpanTraceCompletionTime() throws RetryAttemptException {
...
+        Mockito.when(spanCacheMock.getTrace(null, "trace"))
+                .thenReturn(new HashSet<>(Arrays.asList(rootSpan, descendant)));
 
         completionTimeProcessing = new CompletionTimeProcessing(rootSpan);
         completionTimeProcessing = completionTimeProcessingDeriver.processOneToOne(null, completionTimeProcessing);
         Assert.assertNull(completionTimeProcessing.getCompletionTime());
 
         completionTimeProcessing = completionTimeProcessingDeriver.processOneToOne(null, completionTimeProcessing);
-        Assert.assertEquals(2, completionTimeProcessing.getCompletionTime().getDuration());
+        Assert.assertEquals(2000, completionTimeProcessing.getCompletionTime().getDuration());
```
pavolloffay: In this case it is better to change completion time, as the Span units were not changed.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceServiceITest.java
pavolloffay: Time units on criteria were not changed. Either I change time units on trace objects or on here. 

-> I have changed time values on trace objects.

server/elasticsearch/src/main/resources/hawkular-apm-mapping.json
pavolloffay: Do we keep it as it is now and in the next PR bring it back with some tests that it works? Is it ok for you?

objectiser: Yes that is fine - but can you create the jira for it and include ref to it here.

pavolloffay:
About
`TimeUnit.NANOSECONDS.toMicros(System.nanoTime())` vs `TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis())`

when using nanos to micro in integration tests (`TraceServieITest/AnalyticsServiceITest/...`) tests were failing because `TimeUnit.NANOSECONDS.toMicros(System.nanoTime())` returned `510359034483`

objectiser:
Maybe tests are ok - it was mainly in the actual code where maintaining some of the precision of nanoseconds is important.

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/ElasticsearchUtil.java
```diff
@@ -47,11 +49,14 @@
...
-        long endTime = criteria.calculateEndTime();
-
+        /**
+         * Internally all time units are stored in microseconds
+         * Criteria API accepts milliseconds, therefore range adjustment is needed
+         */
         BoolQueryBuilder query = QueryBuilders.boolQuery()
-                .must(QueryBuilders.rangeQuery(timeProperty).from(startTime).to(endTime));
+                .must(QueryBuilders.rangeQuery(timeProperty)
+                        .from(TimeUnit.MILLISECONDS.toMicros(criteria.calculateStartTime()))
```
objectiser: We need to check about this - if the timestamp being stored with records in the ES db is using its own timestamp (as we don't specify the mapping to our field), then it would be in milliseconds I guess - so the query should take milliseconds?
The docs aren't clear, so may need to experiment.

server/processors/src/test/java/org/hawkular/apm/server/processor/tracecompletiontime/TraceCompletionInformationInitiatorTest.java
```diff
@@ -141,7 +141,7 @@ public void testProcessSingleComponentNoProducers() {
         assertEquals(trace.getBusinessTransaction(), ci.getCompletionTime().getBusinessTransaction());
         assertEquals(trace.getStartTime(), ci.getCompletionTime().getTimestamp());
         assertEquals(EndpointUtil.encodeClientURI(c.getUri()), ci.getCompletionTime().getUri());
-        assertEquals(200, ci.getCompletionTime().getDuration());
+        assertEquals(200000000, ci.getCompletionTime().getDuration());
```
objectiser: Wondering why this one has 6 more zeros?

tests/dist/src/test/java/org/hawkular/apm/tests/dist/AnalyticsServiceITest.java
```diff
@@ -1159,7 +1165,7 @@ public void testGetCompletionTimeMultiFragment() throws Exception {
         assertNotNull(stats);
         assertEquals(1, stats.size());
 
-        assertEquals(1750, stats.get(0).getAverage());
+        assertEquals(1749999, stats.get(0).getAverage());
```
objectiser: This still looks wrong - should be 1750000, unless its an accuracy issue in ES?
If you can't see anything obvious, create a jira and assign to me, to check after this has been merged.

tests/dist/src/test/java/org/hawkular/apm/tests/dist/TraceServiceITest.java
```diff
@@ -440,7 +445,7 @@ public void testStoreAndQueryPropertiesExclude() {
     public void testStoreAndQueryCorrelationsInclude() {
         Trace trace1 = new Trace();
         trace1.setId("1");
-        trace1.setStartTime(System.currentTimeMillis() - 4000); // Within last hour
+        trace1.setStartTime(TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()) - 4000);
```
objectiser: 4000 -> FOUR_MS_IN_MICRO_SEC

server/processors/src/test/java/org/hawkular/apm/server/processor/tracecompletiontime/TraceCompletionInformationInitiatorTest.java
pavolloffay: Source data wasn't changed :+1:

server/elasticsearch/src/main/java/org/hawkular/apm/server/elasticsearch/ElasticsearchUtil.java
objectiser: Ah yes, I was probably getting too caught up with the timestamp mapping issue. So might not be a problem after all :)

tests/dist/src/test/java/org/hawkular/apm/tests/dist/AnalyticsServiceITest.java
pavolloffay: https://issues.jboss.org/browse/HWKAPM-709

objectiser:
@pavolloffay review complete, itests all run fine and test run worked fine. Thanks!

pavolloffay:
@objectiser thanks for the review :+1:

PR #665, opened 2016-11-03
HWKAPM-703 : Group properties by name in Instance Table

ammendonca:
image

### DIFF
--- a/ui/src/main/scripts/plugins/e2e/html/details-modal.html
+++ b/ui/src/main/scripts/plugins/e2e/html/details-modal.html
@@ -65,7 +65,7 @@
             <td class="inst-details-timestamp">{{ tData.timestamp | date:'medium' }}</td>
             <!--<td>{{ tData.uri }}</td>-->
             <td class="inst-details-properties">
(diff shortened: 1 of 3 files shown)


objectiser:
@ammendonca Had a problem with very long property values, as the property does not wrap. When I then hovered over the property the tooltip just flashs up on the left side of the screen as a column which seems to go off the page. I will send pic and some data via email.

ammendonca:
@objectiser I was able to (partially.. for me the tooltip showed) replicate and the best I could come with was this:

Unexpanded property and first property (worst case scenario):
image

Expanded property and first property:
image

Unexpanded property and last property:
image

Expanded property and last property:
image

WDYT ? In all of them the tooltip gets a bit cut off, but since the tooltip works more as a preview, it's not that bad.

objectiser:
@ammendonca This is probably the best solution, as such properties are not ideal anyway - but atleast we are making the best attempt to show what we can. Do you want me to review again now?

ammendonca:
@objectiser, meanwhile, playing a bit more with the possibilities, managed to get it a bit better.. worst case is still being cut (a bit less, tho) but other scenarios are showing fully, eg:

image

image

I'll be pushing the changes now.

objectiser:
@ammendonca Works well, thanks!

PR #668, opened 2016-11-07
HWKAPM-602 : Create hkTime filter for durations

ammendonca:
Add an angular filter to convert from μs duration to human readable format.

image

### DIFF
--- a/ui/src/main/scripts/plugins/e2e/ts/e2ePlugin.ts
+++ b/ui/src/main/scripts/plugins/e2e/ts/e2ePlugin.ts
@@ -26,5 +26,44 @@ module E2E {
     // Doesn't link an empty block
   }]);
 
+  _module.filter('hkTime', function (limitToFilter, numberFilter) {
+    let milliRange =  1000;
+    let secondRange = 1000 * milliRange;
+    let minuteRange =   60 * secondRange;
+    let hourRange =     60 * minuteRange;
+    let dayRange =      24 * hourRange;
+    let weekRange =      7 * dayRange;
+    let yearRange =    365 * dayRange;
+
+    return function (input) {
+      if (input && !isNaN(input)) {
+        if (input > yearRange) {
+          let weeks = input / weekRange;
+          return Math.floor(weeks / 52) + 'y ' + numberFilter(Math.floor(weeks % 52), 0) + 'w';
+        } else if (input > weekRange) {
+          let days = input / dayRange;
+          return Math.floor(days / 7) + 'w ' + numberFilter(Math.floor(days % 7), 0) + 'd';
+        } else if (input > dayRange) {
+          let hours = input / hourRange;
+          return Math.floor(hours / 24) + 'd ' + numberFilter(Math.floor(hours % 24), 0) + 'h';
+        } else if (input > hourRange) {
+          let minutes = input / minuteRange;
+          return Math.floor(minutes / 60) + 'h ' + numberFilter(Math.floor(minutes % 60), 0) + 'min';
+        } else if (input > minuteRange) {
+          let seconds = input / secondRange;
+          return Math.floor(seconds / 60) + 'min ' + numberFilter(Math.floor(seconds % 60), 0) + 's';
+        } else if (input > secondRange) {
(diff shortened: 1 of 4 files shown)


ammendonca:
@objectiser, thoughts ? Also, which other places should we apply it ?

objectiser:
@ammendonca Looks good - just need to change the column label to remove (ms) :)

Other places include:
* Components view - the actual and elapsed columns
* BTxn summary page - the 95% percentile value

Other related areas:
* The min/max filter on the instance table - currently in milliseconds - so this be configurable, or fix on ms for now?
* Duration scale on the charts (components and btxn details pages) - should they be sensitive to the max value?

@pavolloffay Thoughts?

ammendonca:
image

In 95th percentile, the second unit is not displayed, as it breaks the UI with long values (eg: 123ms 456μs)

image

objectiser:
Looks good to me - @jpkrohling @pavolloffay comments?

jpkrohling:
How many chars are needed on the transactions part for the UI to break? If I have 10.000 transactions *per hour* (a very low number), will the UI break?

ammendonca:
@jpkrohling it will take 10.000 transactions properly. even 100.000 and also 1.000.000 (which might need a small adjustment in terms of alignment, but won't break it). The thing is that the units add a lot more space and since they break to a new line (not breaking doesn't help much as it overlaps, see *Booking Service*) the whole card layout gets broken:

image

objectiser:
I think for these cards, just having a single unit may be ok. Also, might want to make the font size for the units smaller?

ui/src/main/scripts/plugins/e2e/ts/e2ePlugin.ts
```diff
@@ -26,5 +26,44 @@ module E2E {
...
+    let hourRange =     60 * minuteRange;
+    let dayRange =      24 * hourRange;
+    let weekRange =      7 * dayRange;
+    let yearRange =    365 * dayRange;
+
+    return function (input, hideSecondary) {
+      if (!isNaN(input)) {
+        if (input > yearRange) {
+          let weeks = input / weekRange;
+          return Math.floor(weeks / 52) + 'y ' + (hideSecondary ? '' : numberFilter(Math.floor(weeks % 52), 0) + 'w');
```
ammendonca: Yes, that was the precision we agreed on at https://issues.jboss.org/browse/HWKAPM-602

ammendonca:
@objectiser, since the filter returns a whole string, making the units text smaller is not trivial and would likely involve mixing html code in the returned output, which I think is not very good practice. For that we'd likely want a directive instead.

I think for now it's ok with a single unit and same size, and in the future we can evaluate other options. If you agree, it's ready to be reviewed as it is.

objectiser:
@ammendonca Will start review, but looks like there is a lint error:

[ERROR] Error: Failed to lint: e2e/ts/e2ePlugin.ts[51, 1]: Exceeds maximum line length of 120, e2e/ts/e2ePlugin.ts[54, 1]: Exceeds maximum line length of 120, e2e/ts/e2ePlugin.ts[56, 1]: Exceeds maximum line length of 120, e2e/ts/e2ePlugin.ts[58, 1]: Exceeds maximum line length of 120.

ammendonca:
@objectiser, oops, will fix.

objectiser:
@ammendonca Just tried running the vertx-opentracing example with ./order.sh joe laptop 1.

When I expanded the instance diagram, the table row at the top did not span the complete length of the diagram. Don't think this is related to this PR - possibly the properties one? So should I create a jira?

Also, one other place we could use the duration unit directive, on the distributed tracing and instance diagram nodes? Again, this could be a separate jira? or would you want to try it in this PR?

objectiser:
@ammendonca Have created https://issues.jboss.org/browse/HWKAPM-730 for the table width issue.
And https://issues.jboss.org/browse/HWKAPM-731 for discussing whether the same duration unit approach should be used on the diagram nodes.

objectiser:
@ammendonca Thanks!

PR #720, opened 2016-12-16
Hawkular Commons was released.

jpkrohling:


jmazzitelli:
the bump to commons is also in PR #718

jpkrohling:
Thanks @jmazzitelli , I just merged that one.

PR #721, opened 2016-12-20
HWKAPM-785 : Highlight Instance nodes with faults

ammendonca:
Show a warning sign on the bottom right for nodes that have a fault property.

image

(Sample image has all nodes with the marker, for viewing how it looks in the different node types)

### DIFF
--- a/ui/src/main/scripts/plugins/directives/instance-view-diagram/ts/instanceViewDiagramDirective.ts
+++ b/ui/src/main/scripts/plugins/directives/instance-view-diagram/ts/instanceViewDiagramDirective.ts
@@ -194,9 +194,13 @@ module InstanceViewDiagram {
           label += '<div class="name">' + uri + '</div>';
 
           let nodeTT = '<ul>';
+          let hasFault = false;
           _.each(d.properties, (property) => {
             if (!!property.value) {
               nodeTT += ('<li class=\'tt-prop\'><strong>' + property.name + '</strong> ' + property.value + '</li>');
+              if (property.name === 'fault') {
+                hasFault = true;
+              }
             }
           });
           nodeTT += '</ul>';
@@ -212,6 +216,11 @@ module InstanceViewDiagram {
             label += '    ' + hkDurationFilter(d.duration);
             label += '  </span>';
             label += '</span>';
+            if (hasFault) {
+              label += '<span class="pull-right fault-indicator">';
+              label += '  <i class="fa fa-warning"></i>';
+              label += '</span>';
+            }
           }
           label += '</div>';
           html += label;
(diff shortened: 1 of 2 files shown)


objectiser:
I am wondering if it needs to be in red to show up more? Could you check with juca and pavol to see what they think. And eventually merge.

ammendonca:
@objectiser, I thought about coloring it for catching more attention, but at the same time I think it makes things a bit too colorful and the red color doesn't fit very well with the background grey:

image
Full [Patternfly red pallete](https://www.patternfly.org/styles/color-palette/#_) applied.

IMO, the user will know what to look for, so it might not be necessary to make it show up a lot more. Maybe make the indicator grey a bit more contrasting with the background, would suffice, here's a sample with several PF greys:

image

CC @cardosogabriel, @jpkrohling, @pavolloffay thoughts ?

jpkrohling:
I prefer the one in the "InventoryDB[QueryInventory]" box, grey. Or is there a meaning for the different grey shades?

objectiser:
+1

pavolloffay:
+1 InventoryDB[QueryInventory] or https://github.com/hawkular/hawkular-apm/pull/721#issue-196784974

ammendonca:
Thanks all for the feedback, there's no different shades of grey, we'll use just a single one. So we'll go with the one from InventoryDB[QueryInventory], it's pf-black-100 (`#fafafa`)

ui/src/main/scripts/plugins/directives/instance-view-diagram/ts/instanceViewDiagramDirective.ts
```diff
@@ -194,9 +194,13 @@ module InstanceViewDiagram {
           label += '<div class="name">' + uri + '</div>';
 
           let nodeTT = '<ul>';
+          let hasFault = false;
           _.each(d.properties, (property) => {
             if (!!property.value) {
               nodeTT += ('<li class=\'tt-prop\'><strong>' + property.name + '</strong> ' + property.value + '</li>');
+              if (property.name === 'fault') {
```
pavolloffay: I know we are using `fault` and `fault.[code|description]` for errors, but OT defines `error` tag
http://opentracing.io/documentation/pages/api/data-conventions.html#errors

we should discuss this when @objectiser is back. (It can be handled in separate jira)

pavolloffay: https://issues.jboss.org/browse/HWKAPM-796

objectiser:
@ammendonca Thanks!

@pavolloffay Have merged as is - but once the error/fault issue has been sorted we can update as necessary.

PR #722, opened 2016-12-21
HWKAPM-793 | bump opentracing version to 0.20.4

pavolloffay:
https://issues.jboss.org/browse/HWKAPM-793

### DIFF
--- a/client/opentracing/src/main/java/io/opentracing/impl/APMSpanBuilder.java
+++ b/client/opentracing/src/main/java/io/opentracing/impl/APMSpanBuilder.java
@@ -14,7 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.opentracing;
+
+package io.opentracing.impl;
 
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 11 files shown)


pavolloffay:
@jpkrohling  could you please review?

client/opentracing/src/main/java/io/opentracing/impl/APMSpanBuilder.java
```diff
@@ -14,7 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.opentracing;
+
+package io.opentracing.impl;
```
jpkrohling: Remind me again why we need to use their namespace, instead of `org.hawkular.apm.opentracing`, for instance

pavolloffay: Because these extend classes from `io.opentracing.impl` with `no` access level modifier.

pavolloffay:
CI failures are not related to this PR (travis fails on UI and jenkins on iptables)

pavolloffay:
@jpkrohling thanks!

PR #730, opened 2017-01-04
HWKAPM-801 | Remove Issue class

pavolloffay:
https://issues.jboss.org/browse/HWKAPM-801

Removing `Issue` class from model.


client/collector/src/main/java/org/hawkular/apm/client/collector/internal/ProcessorManager.java
```diff
@@ -440,7 +440,7 @@ public void process(Trace trace, Node node, Direction direction,
                 // Associate any issues created during initialisation with
                 // the node
                 if (issues != null) {
-                    node.getIssues().addAll(issues);
+                    node.getProperties().add(new Property(Constants.PROP_FAULT, issues));
```
objectiser: These should also be removed - not reported as a property. For now just add local logging.

client/collector/src/main/java/org/hawkular/apm/client/collector/internal/ProcessorManager.java
```diff
@@ -456,7 +456,7 @@ public void process(Trace trace, Node node, Direction direction,
                         pi.setProcessor(processor.getDescription());
                         pi.setSeverity(Severity.Error);
                         pi.setDescription(t.getMessage());
-                        node.getIssues().add(pi);
+                        node.getProperties().add(new Property(Constants.PROP_FAULT, pi));
```
objectiser: Same as above.

client/collector/src/main/java/org/hawkular/apm/client/collector/internal/ProcessorManager.java
```diff
@@ -571,7 +571,7 @@ public void process(Trace trace, Node node, Direction direction,
                 pi.setAction(actionDescription);
                 pi.setSeverity(Severity.Error);
                 pi.setDescription(t.getMessage());
-                node.getIssues().add(pi);
+                node.getProperties().add(new Property(Constants.PROP_FAULT, pi));
```
objectiser: Same as above - ProcessorIssue should also be removed.

pavolloffay: Do we want to remove `ProcessorIssue` altogether?

pavolloffay: sorry I missed the message above :+1: Just ignore my question.

pavolloffay:
What logging level are we going to use in agent? I'm not sure what level is used in all places when there is `ProcessorIssue` created. I saw a couple of `FINE` levels and probably others. Shouldn't we use `SEVERE` in these cases?

objectiser:
Yes I think SEVERE is ok in these cases, as they should rarely happen.

pavolloffay:

api/src/main/java/org/hawkular/apm/api/services/AbstractConfigurationService.java
```diff
@@ -59,20 +57,6 @@
 
                 if (handler != null) {
                     handler.init(processor);
-
-                    if (handler.getIssues() != null) {
-                        for (Issue issue : handler.getIssues()) {
```
objectiser: This part is still required - we need a way for the handler.init method to be able to record issues somewhere so that they can be returned as ConfigMessages from the REST serivce to the UI which will display the errors. So it doesn't need to be done using ProcessorIssue or Issue class - but some way (possibly even ConfigMessage itself) to get the errors from the init method.

pavolloffay:
PR updated

pavolloffay:
I did a mistake while rebasing, please do not review yet.

pavolloffay:
Now it should be fixed.

objectiser:
retest this please

objectiser:
@pavolloffay LGTM :+1:  Integration tests build fine locally - just need to see if jenkins works

pavolloffay:
CI failure is not related to this change (instrumentation-framework ot-polyglot)

pavolloffay:
@objectiser thanks!

Pavol Loffay
Cell SK:    +421 948 286 055
Email: p.loffay@gmail.com

PR #753, opened 2017-01-24
exclude Makefile in maven license-maven-plugin task

rtoma:
This allows me to have a local Makefile for building outside of an IDE.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -591,6 +591,7 @@
         <artifactId>license-maven-plugin</artifactId>
         <configuration>
           <excludes combine.children="append">
+            <exclude>Makefile</exclude>
             <exclude>**/*.data</exclude>
             <exclude>**/*.deployed</exclude>
             <exclude>**/.babelrc</exclude>


jpkrohling:
@rtoma , I'm not quite sure I understand the purpose. How come you have a Makefile on APM's root directory? Are you cloning the repository and adding your build artifacts on top of it? In any case, you can also skip it via a command line argument, IIRC. 

In any case, this brings no bad effects on APM, so, I'm OK with this change. I'd just like to understand why you need it.

rtoma:
As I only want to build the opentracing agent and copy it to a testbed I use this Makefile:
```
all:
	mvn -pl client/agent-opentracing -am install

deploy:
	cp client/agent-opentracing/target/hawkular-apm-agent-opentracing.jar ../hawkular-testing/agent.jar
```

pavolloffay:
Does it make sense to exclude license check on files which are not in repository? 

Why not to use `mvn install -Dlicense.skip=true` for your local testing?

rtoma:
Ah, I've looked for a way to skip the check, but could not find it, so I added the exclude. Thanks for helping me out 👍. Closing this one.

pavolloffay:
If you need anything we are happy to help. You can find us on freenode#hawkular

PR #772, opened 2017-02-17
HWKAPM-860 : Services UI support for "Any" buildstamp

ammendonca:
When selecting the "Any" version, all the  named versions are shown, and
new ones will automatically show as well.

### DIFF
--- a/ui/src/main/scripts/plugins/services/html/services.html
+++ b/ui/src/main/scripts/plugins/services/html/services.html
@@ -18,7 +18,7 @@
             <label class="control-label col-md-3" for="rootnode">Version</label>
             <div class="col-md-9">
               <select pf-select="{noneSelectedText: 'No Versions Available' }" class="form-control" id="buildstamp-select"
-                ng-options="buildStamp.value for buildStamp in buildStamps" ng-model="buildStamp">
+                ng-options="(buildStamp.label || buildStamp.value) for buildStamp in buildStamps" ng-model="buildStamp">
                 <option value="">All Versions</option>
               </select>
             </div>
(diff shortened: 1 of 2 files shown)


ammendonca:
Example of any being used:
image

I have a question, should "Any" also include unnamed/default version ? My concern is that the usability of this feature will decrease as it makes the chart harder to read, as "All" is the sum of all buildstamps, so will always overlap existing lines, eg:
image

@objectiser, @jpkrohling, @pavolloffay, @theute WDYT ?

theute:
IIUC, I think what you show is what I'd like...
We can select Any + all separately, Any shouldn't include All.

Now, are there better names than any/all ?
If we keep the names, I would personally switch...
"Any" == any version, I'm not interested in details --> "single line/color". 
"All" == show me all versions --> "multiple lines/colors".

But I'm sure it's a matter of taste...

Should it be "combined" and "all" maybe ?

pavolloffay:
What does make sens for me: 

`All`: 
show all versions of given service as separate lines. (and if there is a new version available it should be probably added to the chart)

 `Any`
When I saw it my impression is: would it select randomly some version? Or is it the case when there is no version number available (then I would vote for `undefined`)? 


Would be users interested to see an average of all versions? (shouldn't be this screen used to compare versions? If you insist on it what about
* `All individual` - # if lines = # of versions
* `All average` - one line for all versions

objectiser:
Agree we need clearer terms - individual and consolidated?

jpkrohling:
Other than the naming issue, I think it's a great addition.

theute:
Maybe

1. All separated
2. All combined
3. v1
4. v2
5. v3

pavolloffay:
@jpkrohling agree at least with ES it is possible to do it then why not :+1:

ammendonca:
Updated naming to:

- "All" -> "All (Aggregated)"
- "Any" -> "All (Individually)"

Used this rather combined/separated as the labels won't show the whole thing and the abbreviations are more clear, so it will show "All (Agg)" and "All (Ind)" respectively.

image

objectiser:
@ammendonca Looks good to me - but I'll wait for others to comment.

theute:
In which case would "All (Ind)" appear ? I would assume it would have multiple labels with /v2 /v3...

Other than that, sounds good to me

pavolloffay:
I'm very far from typography expert but maybe lower case would be better?

ammendonca:
@theute, It shows in the grey label (where you can remove the selection), as seen in the last screenshot. In the chart, they do show independently as versions.

theute:
@ammendonca indeed, I missed that and that makes sense. I was looking below the chart.

:+1:

objectiser:
@ammendonca Thanks!

PR #782, opened 2017-02-28
HWKAPM-722 : Add timeline directive

ammendonca:
Timeline directive, including:

- Scrubber (line following mouse and showing timestamp);
- Brush Overview to see whole timeline;
- Drag feature to drag in the main timeline;
- Show events (APM logs) at given timeframe;
- Show TX info on click;
- Zoom to TX span on double-click;
- Collapse / Expand span children;

(had a gif demo, but doesn't seem to be able to attach.. will try later)

### DIFF
--- a/ui/src/main/scripts/bower.json
+++ b/ui/src/main/scripts/bower.json
@@ -38,6 +38,7 @@
     "angular-xeditable": "0.1.10",
     "angular-utils-pagination": "0.9.4",
     "dagre-d3": "0.4.17",
+    "d3-tip": "0.6.7",
     "eonasdan-bootstrap-datetimepicker": "4.17.37"
   },
   "devDependencies": {
@@ -45,7 +46,7 @@
   },
   "resolutions": {
     "angular": "1.5.0",
-    "d3": "3.5.5",
+    "d3": "3.5.17",
     "jquery": "1.9.1",
     "lodash": "3.2.0",
     "moment": "2.12.0"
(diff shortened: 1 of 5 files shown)


ammendonca:
Sample usage of the directive:

```
<!-- the actual timeline -->
<div id="timeline-container" timeline nodes="myNodes"></div>
<!-- the overview to brush and zoom in/out -->
<div id="timeline-overview-container" timeline timeline-overview nodes="myNodes"></div>
```

`selectedSpan` is the scope variable that will hold the currently selected span to display info about.

ammendonca:
Demo with improvements

timeline-demo-final

objectiser:
@ammendonca Great work! Marked as 'do not merge' for now until we sort out the backend.

pat2man:
Any progress on the backend? This feature would be great!

objectiser:
@pat2man Sorry for the delay in responding.

Although this UI is great, we have recently made the decision to use the Jaeger project as our OpenTracing core - which already provides its own timeline view. See http://www.hawkular.org/blog/2017/04/19/hawkular-apm-jaeger.html

Therefore we won't be using this timeline view currently - although in the future, potentially if closer integration is required with the OpenShift UI, we may get a chance to use it then.

pat2man:
@objectiser thanks for the response. Jaeger seems like a solid foundation although the aggregated views of Hawkular APM are really what sold me. Looking forward to see those integrated.

PR #787, opened 2017-04-19
Added Jaeger tracer as an option.

jpkrohling:


jpkrohling:
I needed a screenshot to show the vertx-opentracing data on Jaeger along side with the APM representation of it. Not sure this can/should be merged, but this is the result of my changes to get the screenshot.

pavolloffay:
+1 on screenshot. I have nothing against merging. Actually people would see a direct comparison to Jaeger.

Later, we might consider moving examples to a different place.

jpkrohling:
Adding a "do not merge" tag and screenshot.

image

